PR #2, opened 2016-02-28
urlencode now handles parse_qs encoding correctly.

erikash:
The autobahn websocket protocol.py line 2487 uses parse_qs(query) which always returns a list as a return value.
For example:
urllib.parse.parse_qs('room=test') returns {'room': ['test']}. 
In order to parse it correctly later, doseq for urlencode must be set to True.


erikash:
Hi Andrew,
Did you get a chance to check this pull request?

andrewgodwin:
Sorry, I didn't get an email about this - merging now!

PR #6, opened 2016-04-04
Http2 support

flo-dhalluin:
I am working on H2 support (using hyper-h2 python library). It is still very rough, but just to let you know .. 

What's working : 
- streaming request/response
- ssl (required for any browser interaction with http/2) 

What's not working :
- websockets ( I am a bit puzzled whether these are usefull with h2)
- code is a bit messy ( I copied a large part of HTTPFactory instead of factoring in usefull stuff), any feedback on that part ? 
- missing a lot of testing, I only tested with a hello world sample app, with hyper cli tool and firefox... 
- needs rebasing with current master, I know...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,5 @@ python:
   - "3.5"
 install:
   - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install unittest2; fi
-  - pip install asgiref twisted autobahn
+  - pip install asgiref twisted autobahn h2
 script: if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then python -m unittest2; else python -m unittest; fi
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -5,6 +5,8 @@ import importlib
 from .server import Server
 from .access import AccessLogGenerator
 
+from .http_protocol import HTTPFactory
+from .http2_protocol import H2Factory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


andrewgodwin:
Ah, interesting, I was waiting on the Twisted HTTP/2 support to land and then just do it that way (it'd integrate well with the existing Request code, then) - that was kind of how I was planning to avoid duplicating a lot of the existing code like you've had to do.

The WebSocket question is interesting; they do still need to exist, but I have no idea how one negotiates up to them on a pure HTTP2 connection.

flo-dhalluin:
Oh I see, that makes sense.. I just had an itch to try and learn http2. Do they (twisted ..) have a time frame or it's just speculation at that point ? 

Like you said the websocket is still an open question, there's not much in RFCs, it's unclear whether : 
- websockets are new streams multiplexed on the http2 connection (https://tools.ietf.org/html/draft-hirano-httpbis-websocket-over-http2-01 ), that will make work harder with twisted
- they work as they used to, opening a separate socket/connection 
- something else .. 

We should see how it is implemented in browsers maybe... My original idea was that with http2, stuff like long-polling / SSE will be much more reliable and usable and I won't need websockets, hence this experiment ( the next logical step .. wait for it.. would be a django/channels backed SSE module), but I am no HTTP expert, this is all just for "fun".

andrewgodwin:
Last time I talked to them they were getting pretty close to having it merged; certainly far enough along I was willing to wait.

HTTP2 longpoll/SSE is still not going to be as good as WebSockets for chatty connections, though, so they have to work in concert. I suspect multiplexing them on the same http2 connection will be fine, it's easy enough to separate out logical subprotocols in Twisted (already need to for the separate http2 requests anyway).

meshy:
For reference: https://twistedmatrix.com/trac/ticket/7460

hawkowl:
Twisted's HTTP/2 server support has landed in 16.3 which I released today.

andrewgodwin:
This was superseded by #61.

PR #7, opened 2016-04-07
add support for websocket protocols

fcurella:
this simply lets clients that specify a `Sec-WebSocket-Protocol` connect by replying back with the correct headers, provided that that specified protocol was specified on initialization.

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -222,7 +222,7 @@ class HTTPFactory(http.HTTPFactory):
 
     protocol = HTTPProtocol
 
-    def __init__(self, channel_layer, action_logger=None, timeout=120, websocket_timeout=86400, ping_interval=20):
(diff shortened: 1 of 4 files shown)


daphne/http_protocol.py
```diff
@@ -232,7 +232,7 @@ def __init__(self, channel_layer, action_logger=None, timeout=120, websocket_tim
         # We track all sub-protocols for response channel mapping
         self.reply_protocols = {}
         # Make a factory for WebSocket protocols
-        self.ws_factory = WebSocketFactory(self)
+        self.ws_factory = WebSocketFactory(self, protocols=ws_protocols)
```
andrewgodwin: You're passing in "protocols" to WebSocketFactory here but you haven't changed the constructor in this patch - are you sure it's remembering it right?

fcurella: the `protocols` argument is already inherited from the signature in `autobahn.websocket.protocol.WebSocketServerFactory`

PR #23, opened 2016-06-15
Test that malformed urls don't come back

jpic:


jpic:
I think we can do more here, closing for now.

Issue #10, opened 2016-05-07
Daphne should use Twisted's strports/serverFromString for TLS/ACME support


hawkowl:
This would replace this code (https://github.com/andrewgodwin/daphne/blob/master/daphne/server.py#L46) with a user passing a strports description on the command line, which is fed into http://twistedmatrix.com/documents/current/api/twisted.internet.endpoints.html#serverFromString. For example, to support unix sockets it would be `unix:/var/run/finger`, for TCP it'd be `tcp:8000`, for basic TLS it would be `ssl:443:privateKey=key.pem:certKey=crt.pem`, for txsni (TLS w/ SNI) it would be `txsni:certificates:tcp:443`, and for the automatic, turn-on TLS+SNI, it would be `le:/srv/www/certs:tcp:443` (https://github.com/mithrandi/txacme).

I'm happy to implement this, but I thought I'd file an issue since it's been on my mind for so long :)

andrewgodwin:
Sounds good to me! We should keep the `--unix-socket` arg around for a little bit too and make it raise an error saying what to do instead.

mcallistersean:
Now that #37 has landed in master I guess this should JustWork™

I haven't actually found a combination of txacme/txsni that actually works for me, but using 
`
daphne -e lets:/path/to/certs:tcp:interface=<interface_address>:port=443 channel layer project.asgi:channel_layer
`
actually listens on port 443 and tries to get a certificate (at which point it fails). Don't have the time to debug this right now but would like to hear about some success stories.

mcallistersean:
Hi Ameber,
yes, I'm aware of that.
I have a DNs entry pointed to my development machine, port forwarding is in place and it the daphne server is reachable from the outside using the domain name.
Did you manage to get it working and if yes, which versions?
There are also some problems with python3 in txsni.

mcallistersean:
With recent checkouts of txacme and txsni and some patches for py3 compatibility I got lets: endpoints working. But this has nothing to do with daphne anymore, so I guess this issue can be closed.

andrewgodwin:
I've definitely found ACME to be a little bit picky before when doing my own dev, but good to have confirmation it works. I'll close this, but we should get some docs done about it too.

fuzzy76:
Does this mean that Daphne supports HTTPS or not? I have been searching both documentation and issue queue for something definite.

andrewgodwin:
[...]

PR #32, opened 2016-07-13
Support inherited file descriptors

mcallistersean:
I have tried running daphne both in https://github.com/circus-tent/circus/ and https://github.com/unbit/uwsgi
Both expose a file descriptor that is passed from the parent process to daphne, but daphne needs some tweaks to recognize the file descriptor as such.

Usage in circus is something like:

```
[watcher:daphne]
cmd = daphne --fd $(circus.sockets.daphne) <your.channel:layer>
singleton = True
use_sockets = True

[socket:daphne]
port=...
```
It's more complicated in uwsgi to get it to pass the raw socket.

Let me know what you think and I'll write some docs if this approach seems reasonable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,21 @@ set a bind address and port (defaults to localhost, port 8000)::
     daphne -b 0.0.0.0 -p 8001 django_project.asgi:channel_layer
 
 
+If you intend to run daphne behind a proxy server you can use UNIX
+sockets to communicate between the two::
+
+    daphne -u /tmp/daphne.sock django_project.asgi:channel_layer
+
+
+If daphne is being run inside a process manager such as
+`Circus <https://github.com/circus-tent/circus/>`_ you might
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Seems sensible to me if it's a common pattern for passing sockets over. Daphne currently has a sore lack of documentation but if you could add something to the README, I'll come back and fill the other arguments in too.

mcallistersean:
I added a few examples to the README but haven't described all command line flags as I believe that is better served with proper descriptions of the args in the cli.py file.

As for the common pattern, I only know of 2 wsgi servers that support file descriptors directly
```
chaussette --fd 5
uwsgi --socket fd://5
```
So while chaussette usage is identical to my suggestion uwsgi uses something similar to #10 but I don't think twisted's serverFromString supports `fd:` -like strings.

andrewgodwin:
Thanks a lot!

PR #34, opened 2016-08-04
Implement connection force-close via ping timeout

globophobe:
Logging will show WSDISCONNECT. Hopefully, it isn't important to log the reason. As, that would mean duplicating Autobahn's code in Daphne.


andrewgodwin:
Lovely, thanks so much!

PR #36, opened 2016-08-08
Support for HTTPS(secure) connection

faseelakv123:
Support for HTTPS connection


andrewgodwin:
We're actually going to integrate Twisted's new pluggable listening frontends next week, which will bring both TLS support and Let's Encrypt support among other things automatically, but it means rewriting a bit more than this. I'll leave this open in case that fails.

Lukasa:
This can probably be closed now as part of #37.

andrewgodwin:
Agreed.

Issue #31, opened 2016-07-08
Twisted 16.3 appears to never call onMessage in Websocket protocol


andrewgodwin:
The symptom is that the `websocket.receive` channel never gets anything sent to it when people send in websocket frames.

What's happening as far as Daphne is concerned is that onMessage never gets called on the WebsocketProtocol class. I've tested on Python 2 and 3 and both aren't working. Reverting to 16.2 fixes it, so I've pinned Daphne to that for now in the latest patch release while this is investigated.

andrewgodwin:
Verified that Autobahn itself works fine by running the echo server on 16.3, so it's something in daphne.

AlexejStukov:
The new version of twisted (16.3.1) fixed the issue for me. Maybe this can be closed?

andrewgodwin:
When I get a working dev laptop I will verify and remove the pin if true :)

imaurer:
FYI - Daphne on Twisted 16.2.0 works for us. But fails on 16.3.1 and 16.3.2 for us.

Ping me if you need more details. We are reverting Twisted back to 16.2.0 for now.

andrewgodwin:
I have also verified that Twisted 16.3.2 does not fix this, so leaving the pin in place.

@AlexejStukov Could you say what versions of `autobahn` and `daphne` you have installed? Would like to know how it's working for you.

bpeschier:
Ran into this issue today while picking up a side-project and wondered why I left it broken. Turns out updating dependencies is not as harmless as I thought :-)

Running channels 0.17.2, daphne 0.14.3 and autobahn 0.16.0

Twisted 16.2.0 runs fine, 16.3.0 and above (up to todays current, which is 16.4.0) do not.

bpeschier:
This took some digging.

Twisted 16.3 implemented some changes to the http handling, one of which is shutting down (pausing) the producer when all content is thought to be received.

If my reasoning is correct, we only need to resume the producer after we handled the upgrade in ```process()``` of ```WebRequest```. Seems to work in 16.2 and up.

andrewgodwin:
OK, fix finally released in daphne `0.15.0`.

PR #40, opened 2016-08-28
Tell Twisted to keep producing data after connection upgrade

bpeschier:
Twisted 16.3 changed behaviour when all content is thought to be received. "Offending" code is at https://github.com/twisted/twisted/blob/twisted-16.3.0/twisted/web/http.py#L1841

Checked with Twisted 16.2 and it does not break it.

References #31

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -110,6 +110,10 @@ class WebRequest(http.Request):
                     logger.debug("Connection %s did not get successful WS handshake.", self.reply_channel)
                 del self.factory.reply_protocols[self.reply_channel]
                 self.reply_channel = None
+
+                # Resume the producer so we keep getting data
+                self.channel.resumeProducing()
+
             # Boring old HTTP.
             else:
                 # Sanitize and decode headers, potentially extracting root path


andrewgodwin:
Thanks - I thought @hawkowl said it was fixed upstream somehow but this looks clean enough. Will test and release later if it works.

hawkowl:
@andrewgodwin fixed upstream in Autobahn, but only in the AB Resource, which... I forgot you don't use ;)

bpeschier:
Ah, yes :-)

Looks like they did the same thing: https://github.com/crossbario/autobahn-python/blob/5e7898a78dfaa29dce90495a600eb7c0258b03cb/autobahn/twisted/resource.py#L142

My only worry was if pausing and resuming the producer could miss anything while it was paused, but it seems fine.

hawkowl:
@bpeschier there's no issues with that, the OS buffers it :)

andrewgodwin:
Aha, that's why I couldn't just hit upgrade and have it work. This definitely seems like the approach then.

bpeschier:
@hawkowl Good to know. It was a nice dive into Twisted on a lazy sunday afternoon.

@andrewgodwin It might be a good idea to borrow autobahn's attr-check?

andrewgodwin:
@bpeschier Yes, I'll merge that in now as I test.

hawkowl:
The attr check is me being overly cautious (hi, a AB developer here ;) ), as not all transports support the producer interface. In reality, if they don't, something should explode, but I kind of want to never touch that code again :)

andrewgodwin:
I like nothing more than being overly cautious in networking code :)

Merged it in, seems to work well. Release coming up here in a little bit.

PR #41, opened 2016-08-28
Attach path to http.disconnect

damycra:
Set a path key on http.disconnect so it can be routed like `http.request`
https://github.com/andrewgodwin/channels/issues/303


andrewgodwin:
Lovely!

PR #45, opened 2016-09-17
Verbosity

Krukov:
This changes improve debug info at `verbosity` more than 1


andrewgodwin:
If you don't call `globalLogBeginner` then Twisted memory leaks as it processes things; we'd need to still start it but send the output somewhere instead. This is the commit I did to add it in: https://github.com/Krukov/daphne/commit/da40761b95118da82306974e12843566a90ae433

Krukov:
Ok,  I will learn more about twisted logging and fix it.

Krukov:
I hope it's enough. At verbosity mode twisted use the python standart library observer, otherwise logging into nowhere.

PR #48, opened 2016-09-22
Tidy up setup.py a bit

adamchainz:
* Remove unused import 'sys'
* Github is HTTPS
* Add some trove classifiers based upon Django's

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,21 +1,23 @@
 import os
-import sys
+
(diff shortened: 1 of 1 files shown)


andrewgodwin:
The classifiers aren't quite right - Daphne is not directly died to Django and doesn't depend on it or vice versa.

adamchainz:
Dropped django classifier

PR #57, opened 2016-11-05
Fix #55 -- Optionally parse X-Forwarded-For header

### LINKED ISSUE #55: Use X-Forwarded-For headers to populate client addr
Daphne uses the incoming connection's address and port to determine the `client_addr`, which obviously don't exist when listening to a unix socket. When listening to a unix socket we can assume that there's a reverse proxy in front of Daphne so we can check for "X-Forwarded-For" and "X-Forwarded-Port" headers and log those, if they exist.

Also, I'm not sure it's useful to log client ports. It's almost certainly going to be a high random port in both cases. Looking at other servers for reference, neither apache nor nginx seem to log the client port by deafult.

raphaelm:


daphne/ws_protocol.py
```diff
@@ -54,6 +56,15 @@ def onConnect(self, request):
             else:
                 self.client_addr = None
                 self.server_addr = None
+
+            if self.factory.proxy_forwarded_address_header:
```
hackedd: It looks like this should have been `if self.main_factory.proxy_forward_address_header:`?

I'm getting lots of _WebSocketFactory object has no attribute proxy_forwarded_address_header_ errors with this version of Daphne.

raphaelm: Oops, you're right.

raphaelm: https://github.com/django/daphne/pull/59

PR #60, opened 2016-11-05
Add test for WebSocket connection.

hackedd:
Inspired by #59 I've added a test for the WebSocket connection path in `WebRequest`.


andrewgodwin:
Ooh, lovely.

Issue #30, opened 2016-07-06
Support HTTP/2 ✨


hawkowl:
Twisted 16.3.0 landed today, with HTTP/2 support (thanks @lukasa!). This requires:

a) #10, so that it can negotiate it over TLS
b) Daphne's `Site` adding h2 as a supported protocol.

Lukasa:
Actually, it requires more than that. Note that [Daphne does not use `Site` directly](https://github.com/andrewgodwin/daphne/blob/master/daphne/http_protocol.py#L269), but uses its own `HTTPFactory`. This means that it needs to, as well, consider overriding and implementing Twisted's new `_GenericHTTPProtocol` and `_GenericHTTPProtocolFactory` to use from [here](https://github.com/andrewgodwin/daphne/blob/master/daphne/http_protocol.py#L277): otherwise, the HTTP/2 support will not land.

hawkowl:
Oh, ugh. Totally didn't look at that... okay, it'll require a little bit more!

Lukasa:
And to be clear, point b) requires that the `HTTPFactory` here implement [`IProtocolNegotiationFactory`](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IProtocolNegotiationFactory.html).

andrewgodwin:
Yes, Daphne intercepts the HTTP stack in Twisted at a slightly low level as that seemed to be the only way to get access to all the unprocessed data it needs. Hopefully the existing protocol/factory classes can be extended to match the new interfaces; otherwise, I'm up for moving things around,

JohnDoee:
@andrewgodwin Can you elaborate a bit on why it needs to be so low-level? Autobahn websocket runs fine as a Twisted Resource and I can't seem to see anything in http_protocol.py that'd be inaccessible or impossible to make as a Resource.

andrewgodwin:
@JohnDoee I don't know, it might be entirely possible to write as a Resource - the autobahn code I was working from at the time didn't use them for the handoff and so I stuck with what was given to me.

Lukasa:
I should note that, as I look into this a bit further, the fact that Daphne currently does not (and indeed *can not*) terminate TLS totally removes the ability of Daphne to use HTTP/2. Twisted's HTTP/2 support is HTTPS only.

Lukasa:
Of course, there's really no good reason for Daphne not to support HTTPS: it just doesn't.

Lukasa:
[...]

PR #61, opened 2016-11-18
Refactor for HTTP/2 support.

Lukasa:
This PR refactors the internals of Daphne to enable basic HTTP/2 support. This does not enable any of the features for HTTP/2 defined in the ASGI specification, but simply allows for daphne to serve HTTP/2 requests directly without an intervening terminating proxy.

This patch does not work against the current Twisted: it relies on the fix for [Twisted issue 8909](https://twistedmatrix.com/trac/ticket/8909) landing (available in twisted/twisted#597). It will also never work on versions of Twisted before the one that lands that patch.

This fact, along with the fact that Twisted requires several optional dependencies to handle HTTP/2, raises a series of important questions:

1. Should daphne's minimum Twisted version be bumped to the one that supports HTTP/2 in Daphne? (I'd say yes).
2. Should daphne also install the Twisted `http2` extra?
3. What about the Twisted `tls` extra, also required for HTTP/2 and TLS support?

All things worth considering.

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -275,11 +275,6 @@ class WebRequest(http.Request):
         })
 
 
-class HTTPProtocol(http.HTTPChannel):
-
-    requestFactory = WebRequest
-
-
 class HTTPFactory(http.HTTPFactory):
     """
     Factory which takes care of tracking which protocol
@@ -287,8 +282,6 @@ class HTTPFactory(http.HTTPFactory):
     named response channels, so incoming messages can be
     routed appropriately.
     """
-
-    protocol = HTTPProtocol
 
     def __init__(self, channel_layer, action_logger=None, timeout=120, websocket_timeout=86400, ping_interval=20, ping_timeout=30, ws_protocols=None, root_path="", websocket_connect_timeout=30, proxy_forwarded_address_header=None, proxy_forwarded_port_header=None):
         http.HTTPFactory.__init__(self)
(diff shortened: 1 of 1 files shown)


Lukasa:
Resolves #30.

andrewgodwin:
This is a much smaller patch than I was expecting, which is pretty great.

I'm in favour of bumping to only the supported version of Twisted; there's no reason to stick with the current range as apart from maybe distro packages I don't see a need to support lower.

As for the two extras, what are the potential downsides to pulling these in, apart from a larger installation size?

Lukasa:
The only downsides are that it adds an extra set of constraints to the matrix of required versions. Should be no other issue.

andrewgodwin:
Sounds fine on both counts then. Not sure I can merge this until that Twisted release is out, though, so we can pin the version correctly in setup.py?

Lukasa:
Yeah, I think this is on-hold until we can handle the setup.py appropriately.

Lukasa:
The relevant Twisted patch has been merged to master, so the next Twisted release will be the one with the goodness. That'll be either 16.7 or 17.0.

ekimia:
any update on this?

andrewgodwin:
I haven't come back to look at it yet, but the Twisted release that we need is out, so it's now possible I believe.

SRPlatin:
When will http2 be available?

andrewgodwin:
When I can find time to look at this patch, merge it in, and test everything.

andrewgodwin:
I manually merged this in, and updated Daphne's requirement to Twisted 17.0 or greater. I'm not going to require the `http2` or `tls` extras by default; instead, I'll add in some docs on how to get it all working.

PR #62, opened 2016-12-02
Add chdir command line argument

kviktor:
Similar to [uWSGI's chdir](http://uwsgi-docs.readthedocs.io/en/latest/Options.html#chdir).

(it doesn't actually "change directory" but it has the same effect)

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -114,6 +114,12 @@ class CommandLineInterface(object):
             default=False,
             action='store_true',
         )
+        self.parser.add_argument(
+            '--chdir',
+            dest='chdir',
+            help='chdir to specified directory before apps loading',
+            default=".",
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Why would this be needed?

kviktor:
I've tried running daphe with upstart, to make it work I had to change directory to the project first in the script, it kinda felt kinda weird compared to uWSGI which has the chdir option for this.

andrewgodwin:
You can use a virtualenv and install Daphne and your project in that to not need `chdir`, or it's built into any modern init framework (e.g. here's the upstart option: `http://upstart.ubuntu.com/cookbook/#chdir`)

Given that there's already other options, I'm going to close this for now. Sorry.

PR #73, opened 2017-01-30
Fix #72: Allowing `null` origins for Cordova/PhoneGap WS clients

### LINKED ISSUE #72: Receiving a "WebSocket connection denied: origin 'file://' not allowed" HTTP 400 response when initiating a Websocket connection from a Cordova/PhoneGap app
I'm opening this issue to document what was originally posted to the Django Channels project but after investigation deemed related to Daphne itself:
https://github.com/django/channels/issues/392

In short, when running a Cordova/PhoneGap app on Android, the app runs within an embedded browser in a web view. When initiating a Websocket connection to Daphne, the browser sends an `Origin: file://` header on the initial request, which gets refused by Daphne. 

Investigation uncovered that Autobahn (the underlying Websocket library used by Daphne) refuses the connection because the value `file://` is considered invalid - more precisely, it is considered as if [no origin was provided](https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/protocol.py#L2680).

Autobahn does allow the `Origin: file://` to be accepted, but for that to happen the `allowNullOrigin` protocol configuration option must be set to `True`.

As suggested by @andrewgodwin on the original issue, I'm going to work on a change to Daphne that will enable that flag by default, and pass the `Origin` header value to Channels in the `websocket.connect` message headers so that the application can decide wether or not a specific origin is allowed.

davidmarquis:
By setting the `allowNullOrigin` on the Autobahn Websocket factory, we're effectively allowing connections from clients that do not set the Origin header (which is not [compliant](https://tools.ietf.org/id/draft-abarth-origin-03.html#ua-behavior)), or set the `Origin` header value to `file://`, which is the case for at least the Android Web browser when embedded in a Cordova/PhoneGap application.

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -304,7 +304,10 @@ class HTTPFactory(http.HTTPFactory):
         self.reply_protocols = {}
         # Make a factory for WebSocket protocols
         self.ws_factory = WebSocketFactory(self, protocols=ws_protocols)
-        self.ws_factory.setProtocolOptions(autoPingTimeout=ping_timeout)
+        self.ws_factory.setProtocolOptions(
+            autoPingTimeout=ping_timeout,
+            allowNullOrigin=True,
+        )
         self.ws_factory.protocol = WebSocketProtocol
         self.ws_factory.reply_protocols = self.reply_protocols
         self.root_path = root_path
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks!

PR #19, opened 2016-06-15
Use tox for tests

jpic:
Makes it easier to run the test suite on various platforms with one command, and to reproduce the same build on travis and locally.

Could potentially enable pypy and pypy3 support.

PR #75, opened 2017-01-30
Test against Python 3.4 and multiple Twisted versions

maikhoepfel:
I agreed with Andrew what versions of Python the channels project should support, and hence am adding 3.4 here. 

I think it makes sense to support a broad range of Twisted
releases, as users of `daphne` are not unlikely to have other Twisted code
running. It's not feasible to test against all releases since 16.0, and
it would require constant maintenance to add new releases as they come
out. So I opted to keep things simple for now, and only test against the
oldest supported and the current Twisted release.

I'm also adding tox to be able to test against those configurations.

This PR builds on the good work of @jpic in https://github.com/django/daphne/pull/19.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 __pycache__
 dist/
 build/
+/.tox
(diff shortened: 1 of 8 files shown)


andrewgodwin:
:+1:

PR #91, opened 2017-03-07
Full test suite for HTTP requests

maikhoepfel:
The main goal of this pull request is to add a couple tests for HTTP requests. All channel messages generated by those tests also get verified by a new method, which is a translation of the ASGI spec. Instead of using static test data, the tests employ [hypothesis](https://github.com/HypothesisWorks/hypothesis-python), which is rather good at generating data that breaks tests.

As part of this pull requests, some of the existing tests were removed, or moved into a new module. I also stumbled across two minor byte string issues that I could fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ __pycache__
 dist/
 build/
 /.tox
+.hypothesis
+.cache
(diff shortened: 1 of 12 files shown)


maikhoepfel:
I proposed some changes to the ASGI spec in https://github.com/django/channels/pull/554, and updated this pull request with matching tests. 
Travis had a slow day, so it triggered a warning from hypothesis about a slow test. I added a flag to mute the warning for this test, because we expect this test to be slow.

andrewgodwin:
Ugh, somehow I have missed looking at this for over a week. It looks good, I'll pull it in.

PR #92, opened 2017-03-14
Store endpoint listen results.

proofit404:
This little PR allows to take result of server port binding.

This is used in the ChannelLiveServerTestCase where we bind 0 port and than take it result after actual bind.

See https://github.com/django/channels/pull/497 for more details.

### DIFF
--- a/daphne/server.py
+++ b/daphne/server.py
@@ -50,6 +50,7 @@ class Server(object):
         if len(self.endpoints) == 0:
             raise UserWarning("No endpoints. This server will not listen on anything.")
 
+        self.ports = []
         self.signal_handlers = signal_handlers
         self.action_logger = action_logger
         self.http_timeout = http_timeout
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Have you tried to see what happens here if you pass a unix socket rather than a TCP socket to listen? I don't know what you would be appending in that case.

proofit404:
All `tcp`, `ssl` and `unix` endpoints will return deferred object. Its result maybe used for port resolution in the case of `tcp` and `ssl` endpoints. I agree that `ports` may be not the greatest name for this attr.

andrewgodwin:
Yes, in that case it would be nice to have a better name that shows that it's deferreds.

proofit404:
How about `self.listeners`?

andrewgodwin:
Works for me.

PR #104, opened 2017-04-03
Run tox tests from travis.

greut:
It run the tests on Python 3.6 too.

PS: the twisted-old tests could also (and probably should) be achieved using Requirements-Builder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,8 @@ python:
   - "2.7"
   - "3.4"
   - "3.5"
+  - "3.6"
 
-env:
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Can you leave the twisted-old tests stuff in for now, or replace it in the same PR? It's an important part of making sure things are compatible as they release new stuff.

greut:
@andrewgodwin it's there, but tested through _tox_ instead.

andrewgodwin:
Indeed, I missed that. Thanks!

PR #107, opened 2017-04-19
fix check for x-forwarded-for header for WS connections

tykling:
The conversion of twisted headers in line 29 of utils.py converts headers to a dict with keys and values encoded in utf-8, like
`{b'accept': [b'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'], b'dnt': [b'1'], b'cache-control': [b'no-cache']}`

For WS connections the headers are not converted:
`{'accept': ['text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'], 'dnt': ['1'], 'cache-control': ['no-cache']}`

This means that the checks in line 39 and 51 to see if the address_header_name and port_header_name are present in headers fail for WS connections since address_header_name and port_header_name are utf-8 encoded and the dict keys and values are not.

This means that x-forwarded-for and friends are not respected for websocket connections, meaning that you see the proxy IP instead of the real client ip in logs.

This PR fixes the issue by encoding the headers dict keys+values to utf-8 so we have a level playing field.


tykling:
Works for me now, and all tests pass. 
Long term it might be an idea to standardise headers to either utf8 or not, to avoid faffing about with encode and decode.

andrewgodwin:
Hm, headers are supposed to be treated as byte strings with only ASCII characters, so I'd say even using `utf8` to encode might be too fancy and we should just encode with `ascii`. Twisted is giving it to us as it receives them, I suspect, I don't know why autobahn is converting them up to unicode.

Could you reverse `header_value` so it only spits out byte strings and then change the body of the function to do byte comparisons? (e.g. `b',' in address_value`). I think this would be a better approach.

tykling:
Encoding with ascii might be risky as http headers come from the user, and a single high character in there would cause the code to crash when parsing.
I agree that standardising header_values() to only output byte strings is a good idea though, but I think it should be utf-8 encoded. Thoughts?

andrewgodwin:
I guess I'd want to know what Autobahn is using to decode them into unicode in the first place, and we should just mirror that so the transformation is lossless.

tykling:
uhm, I guess something horrible happened while trying to rebase, closing this and opening a new PR

Issue #108, opened 2017-04-19
Unicode chars in a query string in a websocket connection url raise UnicodeEncodeError


ollamh:
Hi, I've just got a bug related to badly formatted query string during websocket connection attempt.

# Traceback
```
[ERROR 2017-04-19 14:36:01,145 daphne.ws_protocol 82 ws_protocol: MainProcess] Traceback (most recent call last):
  File "/.../virtual/local/lib/python2.7/site-packages/daphne/ws_protocol.py", line 47, in onConnect
    query_string = urlencode(request.params, doseq=True).encode("ascii")
  File "/usr/lib/python2.7/urllib.py", line 1368, in urlencode
    l.append(k + '=' + quote_plus(str(elt)))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-23: ordinal not in range(128)
```
# Reproduce error
1. Run daphne as runserver `python manage.py runserver 0.0.0.0:8000` or ` daphne -b 0.0.0.0 -p 8000 asgi:channel_layer`
2. In Chrome console open development console
```
var socket = new WebSocket('ws://localhost:8000/?token=фффàè')
```
3. In Chrome console error:
```
WebSocket connection to 'ws://localhost:8000/?token=%D1%84%D1%84%D1%84%C3%A0%C3%A8' failed: Error during WebSocket handshake: Unexpected response code: 500
```

andrewgodwin:
I cannot reproduce this, I've just done the same locally and it works perfectly. Could you give me the versions you are running of:

* Python
* Django
* channels
* daphne
* asgiref
* asgi_redis
* Operating System

ollamh:
@andrewgodwin 
* Python 2.7.12
* Django 1.8.13
* channels==1.1.3
* daphne==1.2.0
* asgiref==1.1.1
* asgi-rabbitmq==0.4.1 <-- We're using rabbitmq instead of redis
* asgi-redis==1.3.0
* Ubuntu 16.04.2 LTS

Well, when I use `daphne -b 0.0.0.0 -p 8000 ...`, it doesn't return error, but with `python manage.py runserver` I receive this error in stdout. In any case client receives 500 response.
I tried this as well with python `websocket-client`:
```
In [1]: import websocket
In [2]: c = websocket.create_connection('ws://127.0.0.1:8000/ws?token=фффàè')
WebSocketBadStatusException: Handshake status 500
```
Hope that helps to identify the problem

maikhoepfel:
[...]

PR #111, opened 2017-04-28
Websockets test and unicode fix for Python 2

maikhoepfel:
Similarly to my previous work, I'm adding a new test class to help write tests, refactored some of the existing ones, and added a couple new ones. I also fixed a Python 2 byte string issue that also appeared in the HTTP protocol.

This PR currently has one failing test for Python 2, which rather likely is highlighting https://github.com/django/daphne/issues/108, as it fails for unicode query params. I suggest merging this PR, as it eases reproducing the issue.

### DIFF
--- a/daphne/tests/factories.py
+++ b/daphne/tests/factories.py
@@ -84,6 +84,19 @@ def _build_request(method, path, params=None, headers=None, body=None):
     return request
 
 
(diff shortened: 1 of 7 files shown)


andrewgodwin:
Yup, I think you're right about the test. Will merge it in with that included.

PR #129, opened 2017-07-26
Travis - test on Trusty

adamchainz:
As per [their blog post](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming) they're making it the new default, best to be ahead of the curve.

### DIFF
--- a/daphne/tests/test_http_response.py
+++ b/daphne/tests/test_http_response.py
@@ -7,7 +7,7 @@ from __future__ import unicode_literals
 from unittest import TestCase
 
 from asgiref.inmemory import ChannelLayer
-from hypothesis import given, settings
+from hypothesis import given
 from twisted.test import proto_helpers
 
 from daphne.http_protocol import HTTPFactory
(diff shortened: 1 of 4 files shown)


andrewgodwin:
This keeps happening with the hypothesis tests, and it seems to be sticky by platform/python version. I'd suggest adding the settings override that tells it not to do health checks - look in the existing test files to find how, there's 2 examples.

adamchainz:
Added some code that should disable the health check. Used [docs](https://hypothesis.readthedocs.io/en/latest/healthchecks.html) plus [examples I found on GitHub](https://github.com/search?q=suppress_health_check&ref=opensearch&type=Code)

andrewgodwin:
Ah, if you're doing it globally could you also remove the two individual settings overrides (https://github.com/django/daphne/search?utf8=%E2%9C%93&q=perform_health_check)?

daphne/tests/test_http_response.py
```diff
@@ -72,7 +72,6 @@ def test_headers(self, headers):
         headers=http_strategies.headers(),
         body=http_strategies.http_body(),
     )
-    @settings(perform_health_check=False)
```
adamchainz: this was suppressing all health checks, I've just left us suppressing the 'too slow' one only

adamchainz: N.B. f3b5d854ca309dac07436ab0cea7bc3a12c5061b added it and the commit message implies it was just a slow test

andrewgodwin:
Thanks for those changes - as you can see, the previous overrides were not particularly great :)

PR #134, opened 2017-08-21
Get the right client address

### LINKED ISSUE #86: parse_x_forwarded_for uses wrong IP from X-Forwarded-For
Currently Daphne uses the right-most IP when parsing `X-Forwarded-For` while the original client would be the left-most IP.

See: [Wikipedia](https://en.wikipedia.org/wiki/X-Forwarded-For#Format) and [Twisted implementation](https://github.com/twisted/twisted/blob/e920939343ac5f6795c4509e0fe25c33a2447531/src/twisted/web/http.py#L2361-L2362).

millerf:
Fix #86

### DIFF
--- a/daphne/utils.py
+++ b/daphne/utils.py
@@ -4,7 +4,7 @@ from twisted.web.http_headers import Headers
 def header_value(headers, header_name):
     value = headers[header_name]
     if isinstance(value, list):
-        value = value[0]
+        value = value[-1]
     return value.decode("utf-8")
 
 
@@ -37,7 +37,7 @@ def parse_x_forwarded_for(headers,
         address_value = header_value(headers, address_header_name)
 
         if ',' in address_value:
-            address_value = address_value.split(",")[-1].strip()
+            address_value = address_value.split(",")[0].strip()
 
         result = [address_value, 0]
 
--- a/daphne/utils.py
+++ b/daphne/utils.py
@@ -37,7 +37,7 @@ def parse_x_forwarded_for(headers,
         address_value = header_value(headers, address_header_name)
 
         if ',' in address_value:
-            address_value = address_value.split(",")[0].strip()
+            address_value = address_value.split(",")[-1].strip()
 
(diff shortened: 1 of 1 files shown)


JohnDoee:
You're changing the fix from the correct to the previous incorrect version. Check the sources in #86

millerf:
I just saw that. Sorry.
Problem is that I am still getting some wrong values for the "client"... I am using daphne on top of channels, but it seems like channels just passes daphne's "client".
I am not sure where to look at then...

millerf:
My problem is that I have 2 "local" servers in 2 different locations (same city but different places) connecting to a server in Holland. Both "client" sends back the same IP, while [whatismyip.com](https://www.whatismyip.com/) provides 2 distcinct IPs

I just read this on [wikipedia](https://en.wikipedia.org/wiki/X-Forwarded-For#Format):
"Just logging the X-Forwarded-For field is not always enough as the last proxy IP address in a chain is not contained within the X-Forwarded-For field, it is in the actual IP header. A web server should log BOTH the request's source IP address and the X-Forwarded-For field information for completeness."

Could it be that the X-Forwarded-For is not enough to determine the client IP?

andrewgodwin:
`X-Forwarded-For` is not enough if you have a non-cooperating proxy server in the middle. Getting client IPs is not always reliable, unfortauntely.

millerf:
Fir info, on the same server, if using Twisted and retrieve client address with:
self.transport.getPeer().host

I get the right IP. So it is not a problem of proxies... There is definitely something fishy there...

millerf:
@andrewgodwin Any way I could retrieve that IP from the twisted layer in daphne?
It is critical to my project and I just realized that this 'bug' was hapening...

millerf:
Fro info I am only using the websocket part...

millerf:
Nevermind. Just found out my server was not where intended... that caused the problem...
Sorry.

PR #135, opened 2017-08-23
Parse X-Forwarded-Proto header

nicksellen:
We use with [django rest framework](http://www.django-rest-framework.org/) with [django rest swagger](https://github.com/marcgibbons/django-rest-swagger) and noticed that the swagger urls were being generated as `http://` not `https://` (the main site being served as `https://`).

Previously when running with uwsgi server it was working fine.

We found it was because the `X-Forwarded-Proto` was not being honoured in daphne.

Here is the nginx location block we use:

```nginx
location ~ ^\/(api(\-auth)?|docs)\/ {
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
    proxy_set_header Host $host;
    proxy_pass http://unix:/tmp/foodsaving-world-dev.sock;
    proxy_redirect off;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection $connection_upgrade;
}
```

Here's the [uwsgi docs](http://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.8.html#automatically-manage-http-x-forwarded-proto) and [IETF rfc7239](https://tools.ietf.org/html/rfc7239#section-5.4) describing it.

Please make any suggestions for code changes as you feel too :)

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -81,12 +81,18 @@ class WebRequest(http.Request):
                 self.server_addr = None
 
             if self.factory.proxy_forwarded_address_header:
-                self.client_addr = parse_x_forwarded_for(
+                client, proto = parse_x_forwarded_for(
                     self.requestHeaders,
                     self.factory.proxy_forwarded_address_header,
                     self.factory.proxy_forwarded_port_header,
-                    self.client_addr
+                    original=self.client_addr
                 )
+                self.client_addr = client
+                if proto:
+                    # If we get a proto header, force ssl on or off (affects self.isSecure)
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I agree this is needed, but I'd want to see a way to set the header value to read from so it matches the other `X-Forwarded` support (right now, it just turns on if you set the `X-Forwarded-For` header and has a fixed header name it reads from)

nicksellen:
Closed in favour of https://github.com/django/daphne/pull/136

PR #136, opened 2017-08-24
Parse X-Forwarded-Proto header

nicksellen:
Reworked it somewhat from https://github.com/django/daphne/pull/135, in my own fork (thanks for intial work @tiltec!).

I have:
* added tests
* make the header name configurable, as with other similar headers
* reworked the style a bit - to avoid calling `self.setHost()` [1]

How does it look?

[1] it looks like perhaps it _should_ be setting setHost, as per [twisted docs](https://github.com/racker/python-twisted-web/blob/master/twisted/web/http.py#L1129-L1140) but I removed it as it wasn't already setting it

### DIFF
--- a/daphne/utils.py
+++ b/daphne/utils.py
@@ -11,18 +11,22 @@ def header_value(headers, header_name):
 def parse_x_forwarded_for(headers,
                           address_header_name='X-Forwarded-For',
                           port_header_name='X-Forwarded-Port',
-                          original=None):
+                          proto_header_name='X-Forwarded-Proto',
+                          original_addr=None,
+                          original_scheme=None):
     """
     Parses an X-Forwarded-For header and returns a host/port pair as a list.
 
     @param headers: The twisted-style object containing a request's headers
     @param address_header_name: The name of the expected host header
     @param port_header_name: The name of the expected port header
-    @param original: A host/port pair that should be returned if the headers are not in the request
-    @return: A list containing a host (string) as the first entry and a port (int) as the second.
+    @param proto_header_name: The name of the expected protocol header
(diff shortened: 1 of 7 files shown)


daphne/utils.py
```diff
@@ -11,18 +11,22 @@ def header_value(headers, header_name):
...
     Parses an X-Forwarded-For header and returns a host/port pair as a list.
 
     @param headers: The twisted-style object containing a request's headers
     @param address_header_name: The name of the expected host header
     @param port_header_name: The name of the expected port header
-    @param original: A host/port pair that should be returned if the headers are not in the request
-    @return: A list containing a host (string) as the first entry and a port (int) as the second.
+    @param proto_header_name: The name of the expected protocol header
+    @param original_addr: A host/port pair that should be returned if the headers are not in the request
+    @param original_scheme: A scheme hat should be returned if the headers are not in the request
```
tiltec: typo: "hat"

nicksellen: thanks! fixed

andrewgodwin:
Thanks! I'll try to remember to forward-port this into the 2.0 branch (which was forked a few weeks ago), but it's quite different in terms of implementation so I might not get round to it soon.

tiltec:
@andrewgodwin 
Could you (or somebody else) publish a version of daphne that includes this change? Bits of our deployment depend on it.
Thanks!

andrewgodwin:
I just released 1.4.0 with this in it.

MisterFreeze:
a bit broken.
in server.py:101, http_protocol.HTTPFactory is called with proxy_forwarded_proto_header. This, however, is not mentioned in the constructor

MisterFreeze:
Tanks for the update, @andrewgodwin

PR #140, opened 2017-10-12
fix header encoding for ws connections to keep x-forwarded-for functional

tykling:
opened to replace #107


andrewgodwin:
Thanks!

PR #161, opened 2018-02-05
Unix socket fix

### LINKED ISSUE #160: Listen failure: 'UNIXAddress' object has no attribute 'host'
Hello!

I get the above when I try to run daphne 2 using a unix socket. After the error message daphne stops. I did some digging and it seems like the error occurs in daphne/server.py in the listen_success method; getHost() returns a UNIXAddress object which does not have a host or port property.

lideen:
Fixes #160

### DIFF
--- a/daphne/server.py
+++ b/daphne/server.py
@@ -132,7 +132,7 @@ class Server(object):
         """
         if hasattr(port, "getHost"):
             host = port.getHost()
-            if hasattr(host, 'host') and hasattr(host, 'port'):
+            if hasattr(host, "host") and hasattr(host, "port"):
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Looks good - can I get you to fix the linting error and then I'll merge it?

lideen:
Fixed it now, noticed a problem while running the tests on my osx vs. linux OS; the error code for EADDRINUSE was 48 on one and 98 on the other, so I also changed the hardcoded values to use the errno python standard package to avoid the problem, hope that was ok.

andrewgodwin:
That's perfectly fine, and overcomes some laziness on my part. Thanks!

PR #199, opened 2018-05-10
Fixing non-bytes headers, adding proxyhost+proxyport custom headers

karolyi:
The headers on my environment aren't bytes, rather str-s, and so
getting the host and port from those will result None being passed
as a result.

Also, since X-Forwarded-For is not to be trusted, and custom nginx
configurations can pass a `X-Real-IP` header, add two extra command
line parameters to be able to parse custom passed remote IP headers.

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -1,8 +1,7 @@
 import argparse
 import logging
 import sys
-
-from argparse import Namespace, ArgumentError
+from argparse import ArgumentError, Namespace
 from typing import Union
 
 from .access import AccessLogGenerator
@@ -15,6 +14,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_HOST = "127.0.0.1"
 DEFAULT_PORT = 8000
 str_or_none = Union[None, str]
+
 
 class CommandLineInterface(object):
     """
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -2,7 +2,6 @@ import argparse
 import logging
 import sys
 from argparse import ArgumentError, Namespace
-from typing import Union
 
 from .access import AccessLogGenerator
 from .endpoints import build_endpoint_description_strings
@@ -13,7 +12,6 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_HOST = "127.0.0.1"
 DEFAULT_PORT = 8000
-str_or_none = Union[None, str]
 
 
 class CommandLineInterface(object):
@@ -135,7 +133,7 @@ class CommandLineInterface(object):
             default=False,
             action="store_true",
         )
-        self._arg_proxy_host = self.parser.add_argument(
+        self.arg_proxy_host = self.parser.add_argument(
             "--proxy-headers-host",
             dest="proxy_headers_host",
             help="Specify which header will be used for getting the host "
@@ -145,7 +143,7 @@ class CommandLineInterface(object):
             default=False,
             action="store",
         )
-        self._arg_proxy_port = self.parser.add_argument(
+        self.arg_proxy_port = self.parser.add_argument(
             "--proxy-headers-port",
             dest="proxy_headers_port",
             help="Specify which header will be used for getting the port "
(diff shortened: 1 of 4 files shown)


karolyi:
@andrewgodwin, a functionality fix + custom host+port headers passing, at your disposal.
(please merge it, I would like to avoid having to use my own fork.

daphne/cli.py
```diff
@@ -11,6 +13,7 @@
 
 DEFAULT_HOST = "127.0.0.1"
 DEFAULT_PORT = 8000
+str_or_none = Union[None, str]
```
andrewgodwin: Please don't add typing information without updating the entire file (if you want to do that, a separate PR is probably best)

daphne/cli.py
```diff
@@ -132,6 +135,25 @@ def __init__(self):
             default=False,
             action="store_true",
         )
+        self._arg_proxy_host = self.parser.add_argument(
```
andrewgodwin: I don't like these being stored on `self`, I don't think they're needed later

daphne/cli.py
```diff
@@ -146,6 +168,38 @@ def entrypoint(cls):
...
+            message="--proxy-headers has to be passed for this parameter.")
+
+    def _get_forwarded_host(self, args: Namespace) -> str_or_none:
+        """
+        Return the default host header from which the remote hostname/ip
+        will be extracted.
+        """
+        if args.proxy_headers_host:
+            self._check_proxy_headers_passed(
+                argument=self._arg_proxy_host, args=args)
```
andrewgodwin: Don't need to pass the string like this, just redeclare it.

daphne/utils.py
```diff
@@ -15,11 +15,11 @@ def import_by_path(path):
     return target
 
 
-def header_value(headers, header_name):
+def header_value(headers, header_name) -> str:
     value = headers[header_name]
     if isinstance(value, list):
         value = value[0]
-    return value.decode("utf-8")
+    return value.decode("utf-8") if type(value) is bytes else value
```
andrewgodwin: Do you know why your headers are unicode not bytes? Nobody else has ever complained about this, and I don't want to accept this without knowing why.

karolyi: I have no idea why. I can post my rqeuirements.txt if needed, but I run latest-greatest on an ubuntu 16.10 LXC container with python3.6.

daphne/cli.py
karolyi: `self._arg_proxy_host` is not a string, but a parameter that's used for formatting the error messages. If you look at what return value `add_argument` returns, you'll see.

daphne/cli.py
karolyi: The other way is to pass around within return values and parameters. Since the server class is initialized only once per run, I don't see why storing the parameter result for later use is a problem.

daphne/cli.py
karolyi: I use these so that my anaconda on sublime will properly autocomplete. Typing is not enforced on python (yet), so I don't see why this would be a problem, since your requirement is already python3.5+, where this is fully supported.

karolyi:
```
Twisted==18.4.0
txaio==2.10.0
```
```
Python 3.6.3 (default, Oct  6 2017, 08:44:35)
[GCC 5.4.0 20160609] on linux
```

I guess this is as much as I can help you now. I'm on freenode, feel free to contact me over there.

daphne/cli.py
andrewgodwin: It's more that typing information is unrelated to the change as there's not already typing info in the file, so I'd like to keep things consistent and if we add typing do it to the whole file consistently - like if you tried to use `'` quotes not `"`

daphne/cli.py
andrewgodwin: Oh, I know what it is. I'm saying you can just use a string here instead.

andrewgodwin:
That doesn't seem particularly unusual, and I asked a Twisted core dev at PyCon and she told me that Twisted always returns headers as bytes, so I'll need to look into this more.

andrewgodwin:
OK, so I'm happy with the proxy arguments part of the diff, but I can't take the bytes part until I know how to reproduce it locally, as it is apparently impossible. If you want to split the diff up I'll merge the arguments part, but I know that won't solve your direct problem.

I've tried locally with Twisted 17.9 and 18.4, and both were fine. I've merged a more stringent type check into Daphne (b3c097aabdc477ed7e666a4a0d87250db1b2a9d2) to try and see if it uncovers any other people who somehow have this happen to them.

andrewgodwin:
OK, reports of this from at least one other person in the wild: https://github.com/django/daphne/issues/207

Hopefully that will help trace down what's wrong here.

karolyi:
Look at me dancing my I-told-you-so dance :)

andrewgodwin:
Oh I never doubted it was just you - I want to understand why, so it can be fixed lower down.

andrewgodwin:
OK, the bytes issue should be fixed in the main daphne source tree now, so you can back out that part of the diff.

karolyi:
If I do, can you merge it instantly? I use my fork in production and if the bytes issue 'unfixed' version enters there, I'll have a problem.

andrewgodwin:
I mean I can't promise instant merging, I'm currently on a set of flights back home. If the fork is that sensitive, I'd maybe use a second branch and open a new PR?

karolyi:
alright, new PR coming up. have a good flight!

Issue #188, opened 2018-04-06
Run ASGI application constructors in a sync thread


andrewgodwin:
As they may be blocking.

PR #218, opened 2018-07-16
removed deferToThread for ASGI instance constructor

JohnDoee:
As it is not needed anymore (per #188 comment) it should be removed to remove confusion and because it should be unnecessary.

### DIFF
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -3,7 +3,7 @@ import time
 import traceback
 from urllib.parse import unquote
 
-from twisted.internet.defer import inlineCallbacks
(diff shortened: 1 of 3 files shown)


tomchristie:
Refs: https://github.com/encode/uvicorn/issues/129

andrewgodwin:
Can you rebase to see if you can get Travis to run the tests correctly?

JohnDoee:
The error message opaqueness was a case of Twisted being Twisted. Bug fixed and tests look fine now.

andrewgodwin:
Great, thanks!

PR #219, opened 2018-07-22
Add X-Forwarded-Proto support

nicksellen:
From https://github.com/django/daphne/pull/136#issuecomment-324985217

> Thanks! I'll try to remember to forward-port this into the 2.0 branch (which was forked a few weeks ago), but it's quite different in terms of implementation so I might not get round to it soon.

I guess you didn't get around to adding this yet :) 

We just upgraded to channels 2, so I added it again! This time I actually tried running it properly before submitting a PR ;) (the version from my branch is currently being used to serve https://dev.karrot.world).

One extra thing though, I notice it doesn't set the scheme for websocket connections, but it's slightly unclear what the correct behaviour should be so I left it out of this PR for now.

.. but I did try it out on [another branch](https://github.com/nicksellen/daphne/commit/806ddeb74ffbbfcca0ff2b21e3755e6b717aa79f) - however this has slightly odd behaviour as if I use my normal nginx configuration, then it'll end up with _https_ scheme, not _wss_. So options seem to be:
1. do not set "scheme" in scope at all (current behaviour)
2. always set it from the connection (from `self.factory.isSecure`) - ignoring `X-Forwarded-Proto` header
3. map `X-Forwarded-Proto`  value of `https` to `wss` inside `ws_protocol.py`
4. blindly set it from `X-Forwarded-Proto`  header (require user to configure webserver as needed)

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -214,5 +214,6 @@ class CommandLineInterface(object):
             verbosity=args.verbosity,
             proxy_forwarded_address_header="X-Forwarded-For" if args.proxy_headers else None,
             proxy_forwarded_port_header="X-Forwarded-Port" if args.proxy_headers else None,
+            proxy_forwarded_proto_header="X-Forwarded-Proto" if args.proxy_headers else None,
         )
         self.server.run()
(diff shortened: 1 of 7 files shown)


nicksellen:
From a cursory googling session I couldn't find much definitive about how to handle it for the websocket case. I don't currently have any issues with the current implementation, so I would not pursue it further right now.

PR #224, opened 2018-08-24
Run tests against Python 3.7

michael-k:


michael-k:
I just blindly copied `twisted==17.5.0`, but after reading the release notes for daphne 2.2.1, I think think should be updated to `twisted==18.7.0` for all versions of Python.

andrewgodwin:
Please update the python versions list, not adding an extra stage. I agree with the Twisted change though.

michael-k:
It's not (yet) possible to add "3.7" to the list of Python versions; see travis-ci/travis-ci#9815

PR #231, opened 2018-11-07
Configurable servername

Avinash-Raj:
I just want the response header attribute `Server` to display something else like `Apache` instead of `Daphne`. So I made these commits to allow an user to pass their own servername as cli argument.

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -150,6 +150,13 @@ class CommandLineInterface(object):
             "application",
             help="The application to dispatch to as path.to.module:instance.path",
         )
+        self.parser.add_argument(
+            "-s",
+            "--server-name",
+            dest="server_name",
+            help="specify which value should be passed to response header Server attribute",
+            default="Daphne",
+        )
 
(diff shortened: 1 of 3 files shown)


andrewgodwin:
This isn't complete - you aren't overriding it for HTTP connections as well. I'd want to see this complete before I accept it in.

andrewgodwin:
`Server` isn't a required HTTP header last time I checked.

Avinash-Raj:
@andrewgodwin  now the `Server` response header should be attached only if the server name is set to a value other than `Daphne`. But still a testcase fails.  Pls guide me.

andrewgodwin:
Let me re-run that test, it looks flaky.

Avinash-Raj:
@andrewgodwin pls let me know if there is anything that I has to do for getting this PR merged?

andrewgodwin:
My bad, I had just forgotten about this PR (I only get notified about updates, and this didn't get one when I re-ran the check). Merged it now - thanks!

PR #235, opened 2018-11-19
Fixed #234

### LINKED ISSUE #234: When using --fd 0, daphne still listen on 127.0.0.1:8000
When you pass `--fd 0` on the command line, daphne still listens on `127.0.0.1:8000`. It's is due to a bad test on `args.file_descriptor` in `CommandLineInterface.run` method when providing default host and port.

sprat:
Fixed issue #234 where daphne still listen on 127.0.0.1:8000 when --fd 0 is provided


andrewgodwin:
Nice catch - thanks!

PR #237, opened 2018-11-27
Only set disconnected time when it is not already set.

agronick:
This is fix 1 of 2 for django/channels#1181.

### DIFF
--- a/daphne/server.py
+++ b/daphne/server.py
@@ -176,6 +176,9 @@ class Server(object):
 
     def protocol_disconnected(self, protocol):
         # Set its disconnected time (the loops will come and clean it up)
+        # Do not set it if it is already set. Overwriting it might
+        # cause it to never be cleaned up.
+        # See https://github.com/django/channels/issues/1181
         if "disconnected" not in self.connections[protocol]:
             self.connections[protocol]["disconnected"] = time.time()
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
So looking at this change in isolation, I'm curious what effect it would have - did you write it anticipating solving a race condition, or something else? I'd like to see a comment with it because otherwise it looks like you're avoiding replacing a number with another similar-size number which would not really affect memory.

agronick:
I wrote the details in the previous commit:
- check_timeouts() runs every second
- It calls finish() which calls self.server.protocol_disconnected(self) 
- t keeps overwriting the time of the disconnected key
- The disconnected time difference is never high enough to get cleaned up:
```
            if (
                disconnected
                and time.time() - disconnected > self.application_close_timeout
            ):
```
- check_timeouts() continues to be called every second.
- It keeps calling send_disconnect() which causes the memory leak with this code:
```
            self.application_queue.put_nowait({"type": "http.disconnect"})
```
- This makes a giant queue full of useless data.

To experience this just make a view that has time.sleep(999999) with a low http_timeout.

agronick:
I added another fix that handles an exception where send is called on an application instance that has been cleaned up. I see that exception a few times a day.

andrewgodwin:
Great, thanks for adding the comment!

agronick:
Any idea when this is going out? I had this bug hit last night and this is what it did to my memory usage:

image

andrewgodwin:
I'll try to get a release out this week - I forgot to roll one along with Channels last week.

andrewgodwin:
It's now out in `2.2.4`.

agronick:
Great thanks.

PR #278, opened 2019-08-15
Added support for executing from interpreter.

onegreyonewhite:
For run from python interpreter as module use command:
`python -m daphne [all daphne arguments]`.

This is a very useful function for starting the server in integration mode, for example, in uWSGI.
In various Python installations, the directory with `console_scripts` may be different, and in this run option, it is enough to bind to the interpreter.

May be, you could release it feature in patch-release `2.3.1` because this PR does not change functionality but very needed.

### DIFF
--- a/daphne/__main__.py
+++ b/daphne/__main__.py
@@ -1,0 +1,3 @@
+from daphne.cli import CommandLineInterface
+
+CommandLineInterface.entrypoint()


onegreyonewhite:
Guys, this is not such a difficult change. Can you accept and release?

positiveimpact:
This would be super helpful! Currently I'm not able to run daphne on my Azure app service because I can't execute it within the virtual environment in the start script.

carltongibson (approved):
OK, let's add this. Thanks.

PR #327, opened 2020-08-09
Update README for shell compatibility.

mrfredericks:
Some shells aren't too happy when passing arguments that have brackets.  Using quotes around those arguments here should make it easier for folks to copy and paste these commands around. 😄


carltongibson (approved):
OK. I'm a bit Meh, but thanks 🥇

PR #335, opened 2020-10-15
Added request body chunking

### LINKED ISSUE #126: Chunk large request bodies into multiple http.request.body events
I've been searching through the code and Channels seems to support reading chunked bodies, but Daphne always sends the whole body in the `http.request` message.
Am i missing something?

shinuza:
The entire body was previously read in memory which would lead the server to be killed by the scheduler.
This change allows 8 kb chunks to be read until the entire body is consumed.

Fixes #126

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,7 +112,7 @@ should start with a slash, but not end with one; for example::
 Python Support
 --------------
 
-Daphne requires Python 3.5 or later.
+Daphne requires Python 3.6 or later.
 
 
 Contributing
--- a/daphne/http_protocol.py
+++ b/daphne/http_protocol.py
@@ -185,9 +185,19 @@ class WebRequest(http.Request):
                     # Not much we can do, the request is prematurely abandoned.
                     return
                 # Run application against request
-                self.application_queue.put_nowait(
-                    {"type": "http.request", "body": self.content.read()}
-                )
+                buffer_size = self.server.request_buffer_size
+                while True:
+                    chunk = self.content.read(buffer_size)
+                    more_body = not (len(chunk) < buffer_size)
+                    payload = {
+                        "type": "http.request",
+                        "body": chunk,
+                        "more_body": more_body,
+                    }
+                    self.application_queue.put_nowait(payload)
+                    if not more_body:
+                        break
+
         except Exception:
             logger.error(traceback.format_exc())
             self.basic_error(
--- a/daphne/server.py
+++ b/daphne/server.py
@@ -42,6 +42,7 @@ class Server(object):
         signal_handlers=True,
         action_logger=None,
         http_timeout=None,
+        request_buffer_size=8192,
         websocket_timeout=86400,
         websocket_connect_timeout=20,
         ping_interval=20,
@@ -67,6 +68,7 @@ class Server(object):
         self.http_timeout = http_timeout
         self.ping_interval = ping_interval
         self.ping_timeout = ping_timeout
+        self.request_buffer_size = request_buffer_size
         self.proxy_forwarded_address_header = proxy_forwarded_address_header
         self.proxy_forwarded_port_header = proxy_forwarded_port_header
         self.proxy_forwarded_proto_header = proxy_forwarded_proto_header
--- a/daphne/testing.py
+++ b/daphne/testing.py
@@ -18,11 +18,12 @@ class DaphneTestingInstance:
 
     startup_timeout = 2
 
-    def __init__(self, xff=False, http_timeout=None):
+    def __init__(self, xff=False, http_timeout=None, request_buffer_size=None):
         self.xff = xff
         self.http_timeout = http_timeout
         self.host = "127.0.0.1"
         self.lock = multiprocessing.Lock()
+        self.request_buffer_size = request_buffer_size
 
     def __enter__(self):
         # Clear result storage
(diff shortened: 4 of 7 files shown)


carltongibson (commented):
Hi @shinuza Thanks for this. Looks good at first glance. Let me have a play. 🥇

tests/test_http_response.py
```diff
@@ -1,9 +1,8 @@
 # coding: utf8
 
-from hypothesis import given, settings
-
 import http_strategies
 from http_base import DaphneTestCase
+from hypothesis import given, settings
```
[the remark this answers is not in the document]
shinuza: Will do. I'd also like to take a look at #322 to get the build fixed for python 3.5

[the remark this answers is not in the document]
adamchainz: isort 5+ doesn't need special treatment for 3rd party libs?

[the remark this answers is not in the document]
carltongibson: Ok, let me have a play. (Had similar come up the other day on another project and added a config option to get the old result back... 🤔)

carltongibson:
> Will do. I'd also like to take a look at #322 to get the build fixed for python 3.5

Hi @shinuza — Super. That's ≈ready to go. See the similar commit on Channels for how it ended up there https://github.com/django/channels/pull/1458

I'm going to look at this this weekend, so if we can get it ready I'm happy to do a release then, or early next week. 

Thanks for the input! ⛵️

carltongibson:
Hi @shinuza — let's drop Python 3.5 and add Python 3.9. AND drop travis and move to GitHub Actions. See https://github.com/django/channels/pull/1513 for similar on Channels. 

I'm going to focus on the Channels release and then I'll cut back to do a release here. Thanks.

carltongibson:
#336 is updating the CI

shinuza:
> #336 is updating the CI

Should I wait for this to be merged?

carltongibson:
@shinuza — That's done. If you want to rebase... 👍

shinuza:
Just rebased and cleaned the history a bit

README.rst
```diff
@@ -112,7 +112,7 @@ should start with a slash, but not end with one; for example::
 Python Support
 --------------
 
-Daphne requires Python 3.5 or later.
+Daphne requires Python 3.6 or later.
```
adamchainz, line 115: `setup.py` still declares the Python 3.5 classifier, so remove that. It should also have `python_requires=">=3.6"`.

shinuza: This is done already, the 3.5 classifier is not there anymore. :thinking:

daphne/http_protocol.py
```diff
@@ -185,15 +185,31 @@ def process(self):
...
+        """
+
+        chunk = self.content.read(buffer_size)
+        if not chunk:
+            self.application_queue.put_nowait({"type": "http.request", "body": chunk})
+            return
+        incomplete_chunk = len(chunk) < buffer_size
+        payload = {"type": "http.request", "body": chunk, "more_body": False}
+        if not incomplete_chunk:
+            payload["more_body"] = True
```
carltongibson, line 209: Can we combine the `not chunk` and `incomplete_chunk` parts? 

I didn't test, but something like: 

```python
while True: 
    chunk = self.content.read(buffer_size)       
    more_body = not (len(chunk) < buffer_size)
    payload = {"type": "http.request", "body": chunk, "more_body": more_body}
    self.application_queue.put_nowait(payload)
    if not more_body:
        break
```

I'd then be inclined to inline that at line189

shinuza: Ah yeah, I don't like the code so much. This is much better. The only "downside" would be that more_body will still be in the message when there's nothing more to consume, which is not the cleanest way to do things. But I reckon this is still valid per ASGI's spec so that's fine by me.

daphne/http_protocol.py
```diff
@@ -199,16 +199,13 @@ def consume_body(self, buffer_size):
...
-            payload["more_body"] = True
-        self.application_queue.put_nowait(payload)
-        self.consume_body(buffer_size)
+        while True: 
+            chunk = self.content.read(buffer_size)       
+            more_body = not (len(chunk) < buffer_size)
+            payload = {"type": "http.request", "body": chunk, "more_body": more_body}
+            self.application_queue.put_nowait(payload)
+            if not more_body:
+                break
```
carltongibson, line 208: So can we get rid of this extra method and just inline at line 189?

shinuza: Done

daphne/http_protocol.py
shinuza: Done

carltongibson (approved):
OK, thanks for the effort! This looks good to me.

There's just the lint error.

carltongibson:
Grrr. Something has happened to the CI. 🤔 — @michael-k I don't suppose these errors say _"Oh, yes, this…"_ to you do they? Ta.

shinuza:
> Grrr. Something has happened to the CI. thinking — @michael-k I don't suppose these errors say _"Oh, yes, this…"_ to you do they? Ta.

I think's that's on me, I accidentally added my `pyproject.toml` and `poetry.lock`. All good now.

carltongibson:
Ah perfect. 😀

Let's have it.

PR #336, opened 2020-10-19
Switch from Travis CI to GitHub Actions

michael-k:
Based on the work in https://github.com/django/channels by Adam Johnson and Carlton Gibson.

Supersedes PR django/daphne#324.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip wheel setuptools tox
+      - name: Run tox targets for ${{ matrix.python-version }}
+        run: |
+          ENV_PREFIX=$(tr -C -d "0-9" <<< "${{ matrix.python-version }}")
```
michael-k, line 32: Switching to GitHub Actions might be a good opportunity to run the tests on Windows too.  But this line isn't compatible with Windows.

image

Source: https://github.com/michael-k/daphne/actions/runs/316525784

michael-k: A possible solution might be [tox-gh-actions](https://pypi.org/project/tox-gh-actions/), see this commit: https://github.com/michael-k/daphne/commit/49e85f012c03a42d245edecb4e1dbcd14f9e449d.

Tests fail if there are no wheels for twisted, ie. for Python 3.8+ (probably solved in next release, see https://github.com/twisted/twisted/blob/trunk/.github/workflows/wheels.yaml) and for twisted 18.7.0.

carltongibson: Let's bump this to a separate PR. (But yes... 👍)

carltongibson (approved):
This looks good. (Did you get all of Adam's tweaks already? — My quick glance suggests yes.) 

Thanks!

.github/workflows/tests.yml
adamchainz: I moved away from tox-gh-actions in my code since I disliked having an extra set of config to map python versinos to tasks that could rot. Tbh I could turn my shell one liner into a mini tox plugin, or I think it would be a good feature for tox itself e.g. `tox -e pycurrent` could run environments only for the current python version...

setup.cfg
```diff
@@ -1,16 +1,11 @@
...
-
 [flake8]
 exclude = venv/*,tox/*,docs/*,testproject/*,js_client/*,.eggs/*
-ignore = E123,E128,E266,E402,W503,E731,W601
+extend-ignore = E123, E128, E266, E402, W503, E731, W601
 max-line-length = 120
+
+[isort]
+profile = black
+known_first_party = channels,daphne,asgiref,channels_redis,http_base,http_strategies
```
adamchainz, line 8: We dropped `known_first_party` over in channels, if you're modifying the isort setup in this PR might as well copy that all at once.

adamchainz (approved):
Thanks for doing this 👍

.github/workflows/tests.yml
michael-k: > I moved away from tox-gh-actions in my code since I disliked having an extra set of config to map python versinos to tasks that could rot.

Yeah, same here.

> would be a good feature for tox itself

Absolutely!

setup.cfg
michael-k: Sure, will do.

michael-k:
@carltongibson Should be ready; everything's green in my fork: https://github.com/michael-k/daphne/actions/runs/317723477

.github/workflows/tests.yml
michael-k: Btw. over at django-filter I've used Ryan P Kilby's [tox-factor](https://pypi.org/project/tox-factor/) and `python -m tox -f py$(python --version 2>&1 | cut -c 8,10)` which does seem to work on windows.

adamchainz: Great find! Let's try that here?

adamchainz: I also created an issue there to make it possible to use without the shell invocation: https://github.com/rpkilby/tox-factor/issues/14

michael-k: > Let's try that here?

I will, but (as Carlton suggested) in a separate PR.

carltongibson (approved):
OK, super work. Let's have it. Thank you!

PR #341, opened 2020-10-31
Handle asyncio.CancelledError in Server.application_checker

p7g:
As of [bpo-32528](https://bugs.python.org/issue32528) (Python 3.8), asyncio.CancelledError is not a subclass of concurrent.futures.CancelledError.

This means that if an asyncio future raises an exception, it won't be caught. Therefore, the exception will bubble past the try-except within the loop in application_checker, resulting in done applications not being cleaned up, and the application_checker task not being queued again.

In situations of high load, this was causing a very alarming memory leak.

### DIFF
--- a/daphne/server.py
+++ b/daphne/server.py
@@ -277,7 +277,7 @@ class Server(object):
             if application_instance and application_instance.done():
                 try:
                     exception = application_instance.exception()
-                except CancelledError:
+                except (CancelledError, asyncio.CancelledError):
                     # Future cancellation. We can ignore this.
                     pass
                 else:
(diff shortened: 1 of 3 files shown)


adamchainz:
Thanks for spotting this. Ideally this would've been caught by a test that failed on Python 3.8+ - can you check what tests exist and look into adding one?

p7g:
Good call on the test. I had a go at writing one, though it was a little tricky. Do let me know if you have any feedback on how to make it simpler or easier to understand.

carltongibson (approved):
Hi @p7g --- Thanks for this. 

Good catch, and a nice test -- horrible, but nice. 😄 
The fix here is clearly correct. 

I'll put a release together later in the week. 

🥇

PR #351, opened 2021-01-17
Only force a default endpoint w/o valid endpoint

fitzgeralddp:
If you run `daphne --endpoint ssl:8000:privateKey=key.pem:certKey=cert.pem --bind 0.0.0.0  MyApp.asgi:application` in trying to specify that the application should bind to 0.0.0.0, two bindings will be made to the same port, and daphne will never properly start. While I know the `--bind 0.0.0.0` specification is unnecessary, it's certainly confusing to debug and is not intuitive from the standpoint of the programmer trying to use the CLI.

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -245,10 +245,18 @@ class CommandLineInterface(object):
             # no advanced binding options passed, patch in defaults
             args.host = DEFAULT_HOST
             args.port = DEFAULT_PORT
-        elif args.host and args.port is None:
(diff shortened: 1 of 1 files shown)


carltongibson (changes requested):
Hi @fitzgeralddp — you have a syntax error. And no tests. 

Wondering if it would be better to adjust the arg parser to not accept the `bind` in this case 🤔

I'm going to close this since it all the tests are failing, but happy to look at it if you want to update. Thanks.

PR #353, opened 2021-02-03
Added load_asgi_app method to CLI class

makondratev:
In compiled projects(using PyInstaller for example) application must be imported during compilation. So you can redefine only this method (like  gunicorn's load_wsgiapp)  instead of copying whole run() method and rewriting only one line. Hope this will help everyone who compile python projects with daphne server.

### DIFF
--- a/daphne/cli.py
+++ b/daphne/cli.py
@@ -200,6 +200,12 @@ class CommandLineInterface(object):
         if args.proxy_headers:
             return "X-Forwarded-Port"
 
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
Hi @makondratev. This looks good thanks. 

I'll have a look at the equivalent in gunicorn and this about whether we need docs and tests, but seems an easy addition for the next release.

makondratev:
Thanks a lot! That would be great!

PR #387, opened 2021-10-25
Removed deprecated —ws-protocols CLI option.

Kub-AT:
`--ws-protocol` should be removed.


carltongibson (approved):
Thanks for this @Kub-AT — Super.

PR #436, opened 2022-09-01
Add changelog to sdist

jayvdb:


adamchainz:
I recommend not distributing the tests. 99.99% of installs do not need them. Those who want them are normally redistributors, can who fetch the tarball from the repo. Last year I removed tests from the sdists in all my projects and heard no complaints.

Is redistributing your motivation here @jayvdb ? Have you considered using repo tarballs?

jayvdb:
https://pypi.org/project/daphne/#files includes wheels, which is what 99% of installs will fetch.
The sdist is for the other 1% need.

Yes, you broke https://build.opensuse.org/package/show/devel:languages:python:django/python-django-cors-headers which is why it hasnt been updated and is still 3.7 , because the automated update system only works when fetching from pypi.

carltongibson (approved):
OK, I don't mind adding this. (🤷)

Thanks.

PR #446, opened 2022-10-25
Run tests against Python 3.11 and add trove classifier

michael-k:
Python 3.11.0 has been released on 2022-10-24, see https://pythoninsider.blogspot.com/2022/10/python-3110-is-now-available.html

~Python 3.11 is not yet available on GitHub actions.  I'll mark the PR as ready for review once it's available.  See actions/runner-images#6459~
Looks like it was added since I last checked.  ref actions/python-versions#193


adamchainz (approved):
Wicked

PR #450, opened 2022-11-07
Add --nostatic and --insecure args to runserver command. Fixes django/daphne#449

davidfb:
Both arguments are accepted by django.contrib.staticfiles' runserver command, so Daphne's should include them too to provide maximum compatibility.

### DIFF
--- a/daphne/management/commands/runserver.py
+++ b/daphne/management/commands/runserver.py
@@ -73,6 +73,18 @@ class Command(RunserverCommand):
                 "seconds (default: 5)"
             ),
         )
+        parser.add_argument(
+            "--nostatic",
+            action="store_false",
+            dest="use_static_handler",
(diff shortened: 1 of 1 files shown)


carltongibson (commented):
Are these arguments used anywhere? 🤔

davidfb:
Yes. The options populated by these arguments were already in use on Daphne's code (https://github.com/django/daphne/blob/main/daphne/management/commands/runserver.py#L150) but without the arguments definition, those options would never be able to be populated with a value other than their defaults.

carltongibson:
Good. Thanks.

josephetareh:
Any updates on this merge?

Kaju-Bubanja:
Any updates on this merge?

carltongibson:
No, let's have it. Thanks @davidfb

PR #458, opened 2022-12-06
Upgrade to tox 4

adamchainz:
tox 4 final is expected shortly: https://fosstodon.org/@gaborbernat/109428598380006519

tox-py no longer works with it, but that’s okay since the new `-f` factor option fills the need to run Python 3.X tests in the Python 3.X CI run. I documented this pattern in the package install instructions: https://github.com/adamchainz/tox-py/#installation . Plus, I’ve already changed all the projects I maintain to use it, for example: https://github.com/adamchainz/django-htmx/pull/296

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip setuptools wheel
-          python -m pip install --upgrade tox tox-py
+          python -m pip install --upgrade 'tox>=4.0.0rc3'
 
       - name: Run tox targets for ${{ matrix.python-version }}
-        run: tox --py current
+        run: tox run -f py$(echo ${{ matrix.python-version }} | tr -d .)


adamchainz:
tox 4 parses setup.cfg to figure out which dependencies to install - this requires porting from the legacy setup.py to setup.cfg for this to work, plus adding a `pyproject.toml` for PEP 517 package builds.

There’s a bug from tox 4 rc 2+ that means the install of `twisted[tls]` fails - reported upstream: https://github.com/tox-dev/tox/issues/2604 . To get this to pass I pinned to RC1 but we should probably wait before merging.

gaborbernat:
> tox 4 parses setup.cfg to figure out which dependencies to install - this requires porting from the legacy setup.py to setup.cfg for this to work, plus adding a `pyproject.toml` for PEP 517 package builds.

You should not need to do anything as while we do use https://peps.python.org/pep-0621/ we fallback to https://peps.python.org/pep-0517/#prepare-metadata-for-build-wheel when getting the install dependencies. So pure setup.py should also work, just the entire process would be a bit slower.

adamchainz:
You’re right - I think I must have gotten confused whilst tracking down https://github.com/tox-dev/tox/issues/2604 . But now I’ve ported, I think we should keep it, declarative metadata is much nicer :)

carltongibson (approved):
Hey @adamchainz — Super. Thank you for taking this on 🏅 

You're 🟢 when you're ready, presumably when the final release drops, but as you will 😜

PR #509, opened 2024-04-11
Fix package missing management folder(#508)

1One0:
In `setup.cfg`

`packages = find:` was changed to
```
packages =
    daphne    
    twisted.plugins
```

When manually list packages, should enumerate all the packages

```
packages =
    daphne
    daphne.management
    twisted.plugins
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ package_dir =
     twisted=daphne/twisted
 packages =
     daphne
+    daphne.management
     twisted.plugins
 include_package_data = True
 install_requires =


carltongibson:
Thanks for this @1One0 

`commands` is a package too no?

I was wondering if we could fix the `find:` by setting `package_dirs` 🤔

carltongibson:
Screenshot 2024-04-11 at 11 32 32

Built wheel missing the commands package. We could add that as well, but I'd rather fix the `find:`.

1One0:
What about add a extra `daphne.management.commands`?

Or move the dir `/daphne/twisted/plugin` to `/twisted/plugin` then can easily use:
```
package_dir =
packages =find:
```

The problem is hard to change top-level with `setup.cfg`, it will be easier when use `setup.py`.

carltongibson:
Another take on this in #510. 

If you checkout the branch you can `python -m build` and then inspect the wheel to see **both** the twisted plugin and the Daphne subpackages picked up.

1One0:
oh i see, and this problem is exist earlier, it happend when change`find:`
I will try to find a way to fix that...

PR #526, opened 2024-08-18
fix tests with Twisted 24.7.0

dotlambda:
In the fixed test cases the responses now contain `HTTP/1.1` rather than `HTTP/1.0`.

fixes https://github.com/django/daphne/issues/525


carltongibson (approved):
Looks good. Thanks @dotlambda 🎁

PR #529, opened 2024-08-24
Add nightly testing

bigfootjon:


carltongibson (commented):
So… we get a break like in #525 every what 3 or 4 years maybe. I’m wondering if that requires running the tests every day? WDYT?

bigfootjon:
Hm maybe not, I guess we can just fix it when it arises.

PR #539, opened 2024-11-18
Added support for Python 3.13.

smithdc1:
I had one test fail locally but it also came with a load of Windows related errors. So we'll see what the CI has to say instead 😆


carltongibson (approved):
Looks good!

PR #540, opened 2024-11-18
Target py39-plus with pyupgrade.

smithdc1:
I noticed that the minimum Python version is now 3.9. The first commit updates the pyupgrade setting for this.

While here I also ran pre-commit autoupdate -- just the one change, and as it happens that was to pyupgrade as well.


carltongibson (approved):
Great. Thanks

PR #550, opened 2025-02-20
allow assigning a port in DaphneProcess

MohammedSaleh2001:
Needed changes for original idea https://github.com/django/channels/discussions/2135 and ChannelsLiveServerTestCase updates: https://github.com/django/channels/pull/2136#issue-2867092532

Allows DaphneProcess to accept a port. Since I'm not entirely sure what else uses DaphneProcess, I opted to make the parameter optional and at the end of the init function.

### DIFF
--- a/daphne/testing.py
+++ b/daphne/testing.py
@@ -126,7 +126,9 @@ class DaphneProcess(multiprocessing.Process):
     port it ends up listening on back to the parent process.
     """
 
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
OK, thanks for this @MohammedSaleh2001. I'm preparing a Daphne release now. Then we'll cut back to the Channels side of it. 🎁

PR #581, opened 2026-06-04
Add ability to configure websocket message size when running as django dev server

### LINKED ISSUE #580: Now way to extend max message size limit in dev mode
Release 4.2.2 introduces new CLI arguments websocket-max-message-size and websocket-max-frame-size with default value set to 1Mb. However in development mode (when you run application with `manage.py http`) there is no way to set these settings, as result if you application needs 2Mb you can't run it properly locally.

corvis:
Adds ability to configure `websocket-max-message-size` and `websocket-max-frame-size` when running as a django devserver (`manage.py runserver`). Can be configured either via cli argument or django setting. 

Fixes #580

### DIFF
--- a/daphne/management/commands/runserver.py
+++ b/daphne/management/commands/runserver.py
@@ -73,6 +73,22 @@ class Command(RunserverCommand):
                 "seconds (default: 5)"
             ),
         )
+        parser.add_argument(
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
Great. Thanks.

